PR #5, opened 2014-06-13
Updated to use string addresses.

psychoticmeow:
Now fully supports string addresses, you can specify either the `tcp` or `udp` scheme, or leave it empty for backwards compatibility and to allow the executor to choose the best scheme:
- If you specify the `tcp` scheme, only that scheme will be used,
- If you specify the `udp` scheme, it may be switched to `tcp` if the first request fails,
- If you don't specify any scheme then either `tcp` or `udp` will be used depending on the query size.

### DIFF
--- a/src/Query/Executor.php
+++ b/src/Query/Executor.php
@@ -36,9 +36,15 @@ class Executor implements ExecutorInterface
             $scheme = strlen($queryData) > 512 ? 'tcp' : 'udp';
             $nameserver = str_replace('dummy://', null, $nameserver);
             $nameserver = "{$scheme}://{$nameserver}";
+            $keepScheme = false;
         }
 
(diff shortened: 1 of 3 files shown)


clue:
@psychoticmeow Thanks for this PR and your patience, I think this feature makes perfect sense!

Given this is quite old, our APIs have changed significantly in the meantime. In particular, we've removed broken TCP/IP logic (#73) and plan to bring proper TCP/IP support in the future again (#19). With #135 we've made some preparations to properly parse the nameserver URL, so in the future we may use similar logic to explicitly choose an underlying transport.

I'm closing this for now as it hasn't received any input in a while and it's unlikely this will get traction any time soon. The underlying issues referenced above are still open, so I'll make sure to look into this in a future PR. Thank you for your effort and filing this PR!

Issue #13, opened 2014-12-03
Resolving raises an error in the Stream component


clue:
Resolving any name raises an error in the Stream component:

```
Warning: feof(): 60 is not a valid stream resource in .. /src/Socket/Connection.php on line 18
```

This happens because this component uses the `React\Socket\Connection` class instead of the `React\Stream\Stream` class.

Originally reported in: https://github.com/reactphp/socket/issues/4
Documentation bug: https://github.com/reactphp/socket/pull/13
See also: #11

---

> The following is copied from https://github.com/reactphp/socket/issues/4, originally reported by @jeremykendall

I got (almost) the same error this morning while working on the DNS basic usage example. Maybe this could help replicate the warning:

``` php
<?php
// dns.php
error_reporting(PHP_INT_MAX);

require_once 'vendor/autoload.php';

$loop = \React\EventLoop\Factory::create();
$factory = new \React\Dns\Resolver\Factory();
$dns = $factory->create('8.8.8.8', $loop);

$dns->resolve($argv[1])
    ->then(
        function ($ip) { echo "Host: $ip\n"; },
        function ($e) { echo "Error: {$e->getMessage()}\n"; }
);

$loop->run();
```

Running php dns.php jeremykendall.net returned:

``` bash
$ php dns.php jeremykendall.net
Host: 50.57.159.57
PHP Warning:  feof(): 33 is not a valid stream resource in /Users/jeremykendall/dev/explore-react/vendor/react/react/src/Socket/Connection.php on line 18
PHP Stack trace:
PHP   1. {main}() /Users/jeremykendall/dev/explore-react/dns.php:0
PHP   2. React\EventLoop\StreamSelectLoop->run() /Users/jeremykendall/dev/explore-react/dns.php:17
PHP   3. React\EventLoop\StreamSelectLoop->waitForStreamActivity() /Users/jeremykendall/dev/explore-react/vendor/react/react/src/EventLoop/StreamSelectLoop.php:201
PHP   4. call_user_func:{/Users/jeremykendall/dev/explore-react/vendor/react/react/src/EventLoop/StreamSelectLoop.php:227}() /Users/jeremykendall/dev/explore-react/vendor/react/react/src/EventLoop/StreamSelectLoop.php:227
PHP   5. React\Socket\Connection->handleData() /Users/jeremykendall/dev/explore-react/vendor/react/react/src/EventLoop/StreamSelectLoop.php:227
[...]

PR #15, opened 2015-01-05
Replace React\Socket\Connection with client TCP/UDP connection

andig:
This is an _idea_ (and nothing more!) how the fix for https://github.com/reactphp/dns/issues/13 could look like.

This approach needs an additional PR against https://github.com/reactphp/socket-client/blob/master/src/Connector.php to allow creation with `null` Resolver (already allowed for the UDP socket).

The tests don't work due to the mock logic and I was not able to fix them, so not tested at all...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,8 +6,9 @@
     "require": {
         "php": ">=5.4.0",
         "react/cache": "0.4.*",
-        "react/socket": "0.4.*",
-        "react/promise": "~2.0"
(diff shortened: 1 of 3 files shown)


clue:
Thanks for filing this PR! :+1: I believe this has been resolved in the meantime, so I'm closing this for now. Please come back with more details if this problem persists and we can reopen this :+1:

Issue #16, opened 2015-02-01
Fatal on call $conn->close() if connection was not created.


DangerLifter:
Query/Executor::doQuer(). if createConnection throws exception (eg. ErrorException: stream_socket_client(): unable to connect to udp://8.8.8.8:53 (Network is unreachable) if not internet connections) than timeout timer calls $conn->close() on "null object", which causes fatal: PHP Fatal error:  Call to a member function close() on a non-object in /var/www/projects/DomainScrapper/vendor/react/dns/Query/Executor.php on line 62

davidwdan:
This seems like a pretty serious issue to be left unaddressed for over a year.  

I run into this every time that my laptop wakes up and a script tries to reconnect before the network is up.

clue:
For the reference: This bug has actually been addressed somewhat indirectly via https://github.com/reactphp/stream/pull/9 a looong time ago. Ever since this change is in, it should pass through an uncaught exception, but should no longer try to invoke an invalid method on a non-object.

Obviously, this behavior is still broken, so I've just filed #45 that updates the dependencies, handles these exceptions and adds some tests to verify this should no longer happen.

PR #22, opened 2015-03-24
Deal with stream error

DangerLifter:
- correctly proceed stream creation error
- prevent call close() from not existed object (https://github.com/reactphp/dns/issues/16)


cboden:
Can you adjust the formatting to be spaces instead of tabs please?

PR #27, opened 2015-06-02
Add $loop->run(); to actually start the resolving process

ThijsFeryn:
Added $loop->run(); to the documentation. It took me a long time to figure out why the example didn't work.


clue:
Thanks for spotting, that's a nasty one :)

PR #37, opened 2016-03-08
Resolving MX & NS Records

gophry:
Implement message body parsing for types #31 partial implementation. MX AND NS Records.
1. function resolve() in Resolver Class now takes a type argument and defaults to A Record if none is passed.
2. Query for MX and NS records returns an array.
3. For MX queries priority is not captured in the current pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,15 @@ $dns->resolve('igor.io')->then(function ($ip) {
     echo "Host: $ip\n";
 });
 
+$dns->resolve('igor.io',\React\Dns\Model\Message::TYPE_MX)->then(function ($record) {
+    var_dump($record);//returns array of MX Records.
+});
+
+$dns->resolve('igor.io',\React\Dns\Model\Message::TYPE_NS)->then(function ($record) {
+    var_dump($record);//returns array of NS Records.
+});
+
+
 $loop->run();
 ```
 
--- a/src/Protocol/Parser.php
+++ b/src/Protocol/Parser.php
@@ -106,7 +106,6 @@ class Parser
 
     public function parseAnswer(Message $message)
     {
-//        var_dump($message->data);die;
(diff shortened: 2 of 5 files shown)


marcvdm:
I was just trying to figure out why the parser did not give me back all the nameservers. Would be much appreciated if this was to be merged.

gophry:
Feedback appreciated.
nameservers for igor.io, facebook.com, google.com, linode.com work perfectly as sample. cross checked against dig output too.

clue:
Thanks for the PR @gophry, I like this feature, but I'm not so sure about the current interface :+1: 

What does `resolve()` return for `A` records any how this differ for other record types?

gophry:
@clue apologies. i am under medical treatment for some time so was unavailable.
resolve() returns a string for A records and an array for NS and MX.

clue:
No worries, there's no need to hurry these things, I hope you're alright :)

> resolve() returns a string for A records and an array for NS and MX.

What's the reasoning for this? I have to admit I'm not sold on the idea of using a single method that returns a different data structure depending on an input parameter. As an alternative, does it make sense to move this to separate methods instead?

gophry:
I was in sense not comfortable myself with the idea. but the current resolve() returns a string. it will make sense to make them as separate methods.

clue:
FWIW, Node.js also has separate methods for each query type, for example https://nodejs.org/api/dns.html#dns_dns_resolvens_hostname_callback. Its `resolve()` method does in fact resolve with different types of values, I'm not sure this is something we want to do here.

zainengineer:
any estimate on when it will be merged ?

clue:
@gophry Thank you for your patience!

Support for NS records has been merged via #104.
Support for MX record has been merged via #106.

Issue #19, opened 2015-02-23
Support DNS over TCP/IP transport and retry truncated UDP responses


clue:
Resolution via TCP transport should use the [react/socket-client component](https://github.com/reactphp/socket-client).

Fixes #11 and #13.

clue:
Depends on https://github.com/reactphp/socket-client/issues/12 so that the socket-client component does not depend on a DNS `Resolver` instance again.

clue:
Updated title now that the SocketClient component has been deprecated via https://github.com/reactphp/socket/issues/74

clue:
This ticket is ready to be implemented in a future version :+1: 

For the reference: See also the updated [RFC 7766](https://tools.ietf.org/html/rfc7766) and the old [RFC 1035](https://tools.ietf.org/html/rfc1035)

someniatko:
@clue Any estimates, please?

clue:
As much as I'd like to commit to specific date, I can only say that we will get to this feature as soon as time permits.

If you want to sponsor this feature, feel free to reach out :+1:

Issue #12, opened 2014-10-23
Should use datagram component


clue:
Depends on tagged release first (reactphp/datagram#2).

Also refs #11.

PR #47, opened 2017-02-13
Support Socket component v0.5

clue:
Support upcoming Socket v0.5 release. Socket v0.5 contains some breaking changes, but this component is actually fully compatible.

The Socket v0.5 release is *not* out yet, but we'll have to update this component first, because there would otherwise be no installation candidate for when Socket v0.5 tries to install DNS v0.4.* due to its circular dependency.

Ultimately, this dependency should eventually be removed via #12 and #19, but this will require a rather bigger change in the future and a potential BC break.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "require": {
         "php": ">=5.3.0",
         "react/cache": "~0.4.0|~0.3.0",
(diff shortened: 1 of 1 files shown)


clue:
Rebased now that #45 is in :shipit:

PR #48, opened 2017-02-13
Add TimeoutExecutor decorator 

clue:
The `Executor` violates SRP, so this PR is a first step to break this up into multiple independent classes.

This PR preserves full BC by making its timeout feature optional first and then using a `TimeoutExecutor` decorator instead. This can also be reused for other components, in particular clue/mdns-react. Also, this is needed for upcoming changes related to #12 and #19 which will introduce new executor classes that will reuse this timeout handling.

Builds on top of #35 
Somewhat similar to https://github.com/reactphp/socket-client/pull/51

PR #49, opened 2017-02-13
Add simpler internal API

clue:
The `Executor` violates SRP, so this PR is a first step to break this up into multiple independent responsibilities.

This PR preserves full BC by adding some new internal APIs first and then deprecate the existing internal API. This can also be reused for other components, in particular clue/mdns-react. Also, this is needed for upcoming changes related to #12 and #19 which will introduce new executor classes that will reuse these methods.

If you want to review, consider looking at the individual commits instead.

Refs #48.

### DIFF
--- a/src/Model/Message.php
+++ b/src/Model/Message.php
@@ -2,6 +2,8 @@
 
 namespace React\Dns\Model;
 
+use React\Dns\Query\Query;
+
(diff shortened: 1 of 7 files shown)


clue:
Rebased now that #48 is in :shipit:

PR #75, opened 2017-08-16
Support resolving from default hosts file

### LINKED ISSUE #10: Added support for /etc/hosts

clue:
TL;DR: Once this is in, you can now create connections to `localhost` as expected across all platforms (https://github.com/reactphp/socket/issues/88).

This PR adds support for resolving hosts from the system default hosts file. This file is often used for static mappings of certain common hostnames, such as `localhost`.

The hosts file will be loaded once the `Factory` creates the `Resolver` instance. The hosts file is loaded on Unix and Windows based systems alike and if this fails, this will simply be ignored and thus preserves the existing behavior. Windows does not store "localhost" in the hosts file by default, but resolves this internally. This PR also adds static mapping for this, so that the DNS resolver now behaves more like the system resolver.

The hosts file will be preferred for type `A` (IPv4), `AAAA` (IPv6) and `PTR` (reverse DNS) requests. For any other type or if no match is found, it will fall back to the default executor so that this does not affect the existing behavior.

I know this is a pretty big PR, but I've tried to keep the individual commits so reviewing individual changesets may be easier.

Closes #10
Refs https://github.com/reactphp/socket/issues/88

### DIFF
--- a/src/Config/HostsFile.php
+++ b/src/Config/HostsFile.php
@@ -107,4 +107,29 @@ class HostsFile
 
         return $ips;
     }
+
+    /**
+     * Returns all hostnames for the given IPv4 or IPv6 address
+     *
+     * @param string $ip
+     * @return string[]
+     */
+    public function getHostsForIp($ip)
+    {
+        // check binary representation of IP to avoid string case and short notation
+        $ip = @inet_pton($ip);
+
+        $names = array();
+        foreach (preg_split('/\r?\n/', $this->contents) as $line) {
+            $parts = preg_split('/\s+/', $line);
+
+            if (inet_pton(array_shift($parts)) === $ip) {
+                foreach ($parts as $part) {
+                    $names[] = $part;
+                }
+            }
+        }
+
+        return $names;
+    }
 }
--- a/src/Config/HostsFile.php
+++ b/src/Config/HostsFile.php
@@ -12,7 +12,9 @@ use RuntimeException;
  *
  * Most notably, this file usually contains an entry to map "localhost" to the
  * local IP. Windows is a notable exception here, as Windows does not actually
- * include "localhost" in this file by default.
+ * include "localhost" in this file by default. To compensate for this, this
+ * class may explicitly be wrapped in another HostsFile instance which
+ * hard-codes these entries for Windows (see also Factory).
  *
  * This class mostly exists to abstract the parsing/extraction process so this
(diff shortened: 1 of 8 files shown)


src/Config/HostsFile.php
```diff
@@ -0,0 +1,146 @@
...
+        foreach (preg_split('/\r?\n/', $this->contents) as $line) {
+            $parts = preg_split('/\s+/', $line);
+            $addr = array_shift($parts);
+
+            // remove IPv6 zone ID (`fe80::1%lo0` => `fe80:1`)
+            if (strpos($addr, ':') !== false && ($pos = strpos($addr, '%')) !== false) {
+                $addr = substr($addr, 0, $pos);
+            }
+
+            if (@inet_pton($addr) === $ip) {
```
kelunik: But sure, but this might compare `false === false`, no?

clue: If *both* your hosts file is invalid *and* you pass an invalid parameter to this file, then yes. Do you think this should be addressed in a different way?

kelunik: I guess it should just error out if `$ip = @inet_pton($ip)` returns `false`, might help with programming errors.

src/Query/HostsFileExecutor.php
```diff
@@ -76,7 +76,7 @@ private function getIpFromHost($host)
             return inet_ntop(strrev($ip));
         } elseif (substr($host, -9) === '.ip6.arpa') {
             // IPv6: replace dots, reverse nibbles and interpret as hexadecimal string
-            $ip = @inet_ntop(hex2bin(strrev(str_replace('.', '', substr($host, 0, -9)))));
+            $ip = @inet_ntop(pack('H*', strrev(str_replace('.', '', substr($host, 0, -9)))));
```
kelunik: What's the reason for the change from `hex2bin(...)` to `pack('H*', ...)`?

clue: They both do exactly the same thing, except that `hex2bin()` received a number of changes over its lifetime and was introduced later than the `pack()` equivalent, hence why using `pack()` fixes an issue with the last test run :+1:

cboden:
:+1:

src/Config/HostsFile.php
clue: See follow-up PR #76 :+1:

PR #79, opened 2017-09-10
Move react/socket to dev dependency

cboden:
To reduce cyclic dependency between the two, as it is only used in tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "react/stream": "^1.0 || ^0.7 || ^0.6 || ^0.5 || ^0.4.5"
     },
     "require-dev": {
-        "react/socket": "^1.0 || ^0.8 || ^0.7 || ^0.6 || ^0.5 || ^0.4.4",
+        "react/socket": "^1.0 || ^0.8",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -9,10 +9,10 @@
         "react/event-loop": "^1.0 || ^0.5 || ^0.4 || ^0.3.5",
         "react/promise": "^2.1 || ^1.2.1",
         "react/promise-timer": "^1.2",
-        "react/socket": "^1.0 || ^0.8 || ^0.7 || ^0.6 || ^0.5 || ^0.4.4",
         "react/stream": "^1.0 || ^0.7 || ^0.6 || ^0.5 || ^0.4.5"
     },
     "require-dev": {
+        "react/socket": "^1.0 || ^0.8 || ^0.7 || ^0.6 || ^0.5 || ^0.4.4",
```
kelunik: In that case there's no reason to keep the old versions here.

cboden: Good point

cboden:
I wasn't aware of #19. No point in merging this if it's just going to be undone soonish.

PR #88, opened 2017-12-23
Fix test suite forward compatibility with upcoming EventLoop releases

clue:
Refs https://github.com/reactphp/event-loop/pull/102 and https://github.com/reactphp/event-loop/pull/138

### DIFF
--- a/tests/Query/ExecutorTest.php
+++ b/tests/Query/ExecutorTest.php
@@ -29,7 +29,7 @@ class ExecutorTest extends TestCase
     /** @test */
     public function queryShouldCreateUdpRequest()
     {
-        $timer = $this->getMockBuilder('React\EventLoop\Timer\TimerInterface')->getMock();
+        $timer = $this->createTimerMock();
         $this->loop
             ->expects($this->any())
             ->method('addTimer')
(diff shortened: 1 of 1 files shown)


tests/Query/ExecutorTest.php
```diff
@@ -290,6 +291,13 @@ private function createConnectionMock($emitData = true)
         return $conn;
     }
 
+    private function createTimerMock()
+    {
+        return $this->getMockBuilder(
+            'React\EventLoop' . (interface_exists('React\EventLoop\TimerInterface') ? '' : '\Timer') . '\TimerInterface'
```
kelunik: interface_exists('React\EventLoop\TimerInterface') ? 'React\EventLoop\TimerInterface' : 'React\EventLoop\Timer\TimerInterface' would be more readable IMO.

clue: Which is what I used before committing this, but it results in a rather long line :) Any suggestions?

clue: Anyway, updated to also match https://github.com/reactphp/promise-timer/pull/31 :+1:

kelunik: I wouldn't care about a longer line if it's more readable and I don't have to think. Just format it with multiple lines if you care about that:

```php
interface_exists('React\EventLoop\TimerInterface')
    ? 'React\EventLoop\TimerInterface'
    : 'React\EventLoop\Timer\TimerInterface'
```

Apart from that, just drop older PHP versions and use `::class` ;-)

PR #94, opened 2018-02-26
Support parsing DNS config from WMIC on Windows

clue:
This PR implements DNS server config autodetection on Windows by parsing WMIC output. I've looked into reading this from nslookup (works, but blocks when network is not available), ipconfig /all (difficult to parse due to localization), registry (non-trivial implementation) and wmic (which I ended up using). Afaict this command is available on all Windows versions starting with Windows XP and uses some CSV/JSON-based output format that we can parse.

The WMIC command will now be used on Windows by default by the `Config` class as introduced with #93. This PR includes a number of test cases for the expected WMIC output, but I would to see if anybody else can test this on their local Windows setup. Any feedback is much appreciated!

Builds on top of #92 and #93
Refs #29

### DIFF
--- a/src/Config/Config.php
+++ b/src/Config/Config.php
@@ -17,7 +17,7 @@ class Config
      *
      * Note that this method will try to access its files and/or commands and
      * try to parse its output. Currently, this will only parse valid nameserver
-     * entries from `/etc/resolv.conf` and will ignore all other output without
+     * entries from its output and will ignore all other output without
      * complaining.
      *
      * Note that the previous section implies that this may return an empty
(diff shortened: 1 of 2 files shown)


clue:
See also #95 for a follow-up PR to support multiple DNS servers for a single network interface.

I've also had the chance to perform a number of measurements to see how long the WMIC command takes. A usual run takes around 0.15s, with rare (random?) spikes up to 8s on my test server that does nothing in particular. The very first run seems to take ~1.5s.

We *may* want to eventually look into parsing `ipconfig /all` which seems to fluctuate less. Note that parsing likely requires more effort as its output is localized (on some Windows versions?). PRs are much appreciated :+1: 

Also, this feature currently returns a list of *all* DNS servers. It does currently not take "special" DNS servers into account, such as DNS server settings automatically assigned to VirtualBox network interfaces, see https://gist.github.com/clue/ed235ad8d50087d098e4de5fd370b25e#file-gistfile1-txt-L102 and https://www.virtualbox.org/manual/ch09.html#nat-adv-dns.

Practically, this feature is documented to only actually apply the first DNS server from this list (see #93 and also #6). Because phsyical network interfaces seem to always use a smaller "interface ID", this is unlikely to cause any issues for now. PRs to improve this eventually are much appreciated! :+1:

PR #113, opened 2018-10-01
Add support for CAA record types

rabollig:
Reference https://en.wikipedia.org/wiki/List_of_DNS_record_types and
https://en.wikipedia.org/wiki/DNS_Certification_Authority_Authorization


clue:
I'm having to close this for now as it hasn't received any input in a while and it's unlikely this will get traction any time soon. Please comment/update if you think this is still relevant and we can reopen this :+1: 

Thank you for your effort nonetheless, keep it up! :+1:

Issue #114, opened 2018-11-01
PHP Fatal error:  Maximum function nesting level of '256' reached, aborting! Protocol/Parser.php on line 293


wxiaoguang:
When feed invalid data to Parser, we meet the fatal error.

Test data: `hex2bin('80000000016db558016de4c8000000000000000400000001461fffff000005dc000020000000000137cdfa4000000000')`

It is caused by the call between `readLabels` and `getCompressedLabel`
```

PHP 242. React\Dns\Protocol\Parser->readLabels($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:273
PHP 243. React\Dns\Protocol\Parser->getCompressedLabel($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:245
PHP 244. React\Dns\Protocol\Parser->readLabels($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:273
PHP 245. React\Dns\Protocol\Parser->getCompressedLabel($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:245
PHP 246. React\Dns\Protocol\Parser->readLabels($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:273
PHP 247. React\Dns\Protocol\Parser->getCompressedLabel($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:245
PHP 248. React\Dns\Protocol\Parser->readLabels($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:273
PHP 249. React\Dns\Protocol\Parser->getCompressedLabel($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:245
PHP 250. React\Dns\Protocol\Parser->readLabels($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:273
PHP 251. React\Dns\Protocol\Parser->getCompressedLabel($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:245
PHP 252. React\Dns\Protocol\Parser->readLabels($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:273
PHP 253. React\Dns\Protocol\Parser->getCompressedLabel($data = *uninitialized*, $consumed = *uninitialized*) Protocol/Parser.php:245
[...]

PR #116, opened 2018-11-04
Fix interpretation of TTL as UINT32 with most significant bit unset

clue:
TTL is actually a UINT32 with most significant bit unset for BC reasons as per https://tools.ietf.org/html/rfc2181#section-8. This will not affect "normal" operation as the most significant bit should not be set by compliant implementations and we will now interpret this accordingly.

Refs #81
Refs #114

### DIFF
--- a/src/Model/Record.php
+++ b/src/Model/Record.php
@@ -20,7 +20,8 @@ class Record
     public $class;
 
     /**
-     * @var int maximum TTL in seconds (UINT16)
+     * @var int maximum TTL in seconds (UINT32, most significant bit always unset)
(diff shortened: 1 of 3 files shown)


kelunik:
It's probably best to limit the TTL to a much lower value, such as one day, to mitigate the effect of a successful cache poisoning attack.

clue:
@kelunik I think you're raising a valid point when it comes to the caching layer (see #81). However, this PR is concerned with the networking layer. It's my understanding that this layer should therefor implement the DNS protocol specs as suggested in this PR.

kelunik:
@kelunik Indeed, the policy for that shouldn't live in the parser.

PR #121, opened 2019-01-29
Forward compatibility with upcoming Cache v0.6 and Cache v1.0

clue:
Cache v0.6 is going to be tagged soon: https://github.com/reactphp/cache/milestone/4. This component is already compatible with both the upcoming version as well as legacy versions. The Cache component is pretty much stable already and we plan to not introduce any BC breaks anymore: https://github.com/reactphp/cache/issues/8

Builds on top of #102

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "license": "MIT",
     "require": {
         "php": ">=5.3.0",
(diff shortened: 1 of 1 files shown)


kelunik:
What's the reason to do these before the release? I may have already asked this, sorry if that's the case.

kelunik:
@clue thanks!

PR #125, opened 2019-04-01
Cooperatively resolve hosts to avoid running same query concurrently

### LINKED ISSUE #90: Limit number of concurrent DNS queries
Add an executor to limit the number of concurrent DNS queries and queue or reject additional ones.

Additionally, sending the same query in parallel and waiting for a reply for each doesn't make much sense. In particular the caching executor should probably also take this into account.

clue:
This PR adds a new `CoopExecutor` which cooperatively resolves hosts via the given base executor to ensure same query is not run concurrently.

This is useful because all executors are entirely async and as such allow you to execute any number of queries concurrently. You should probably limit the number of concurrent queries in your application or you're very likely going to face rate limitations and bans on the resolver end. For many common applications, you may want to avoid sending the same query multiple times when the first one is still pending, so you will likely want to use this in combination with some other executor (which is the new default).

This is best reproduced by using some higher-level protocol implementation, for example it's rather common to send hundreds of (RESTful) HTTP API calls to the same host name, often concurrently. In this case, it doesn't make much sense to send the same DNS query multiple times.

Note that this is not to be confused with caching (which is already implemented to some degree). This PR only deals with avoiding *concurrency* while caching deals with queries that happen *after* a query is already completed.

Here's a synthetic example (`99-excessive.php reactphp.org 1000`) which highlights this low-level API. Benchmarking suggests this example improved from ~390ms down to ~160ms using a local DNS resolver (YMMV).

```php
<?php

use React\Dns\Config\Config;
use React\Dns\Resolver\Factory;

require __DIR__ . '/../vendor/autoload.php';

$config = Config::loadSystemConfigBlocking();
$server = $config->nameservers ? reset($config->nameservers) : '8.8.8.8';

$factory = new Factory();
$resolver = $factory->create($server, $loop);

if (!isset($argv[2])) {
    echo 'Usage: ' . $argv[0] . ' <domain> <n>' . PHP_EOL;
    exit(1);
}

$domain = $argv[1];
$n = (int)$argv[2];

echo 'go…';
$ok = $failed = 0;
for ($i = 0; $i < $n; ++$i) {
    $resolver->resolve($domain)->then(function () use (&$ok, &$failed, $n) {
        ++$ok;
        echo "\r" . $ok . '/' . $failed . '/' . $n;
    }, function () use (&$ok, &$failed, $n) {
        ++$failed;
        echo "\r" . $ok . '/' . $failed . '/' . $n;
    });
}

$loop->run();

echo "\r" . $ok . '/' . $failed . '/' . $n . ' done'. PHP_EOL;
```

Closes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,26 @@ $executor = new RetryExecutor(
 );
 ```
 
+Note that this executor is entirely async and as such allows you to execute
+any number of queries concurrently. You should probably limit the number of
+concurrent queries in your application or you're very likely going to face
+rate limitations and bans on the resolver end. For many common applications,
+you may want to avoid sending the same query multiple times when the first
+one is still pending, so you will likely want to use this in combination with
+a `CoopExecutor` like this:
+
+```php
+$executor = new CoopExecutor(
+    new RetryExecutor(
+        new TimeoutExecutor(
+            new UdpTransportExecutor($loop),
+            3.0,
+            $loop
+        )
+    )
+);
+```
+
 > Internally, this class uses PHP's UDP sockets and does not take advantage
   of [react/datagram](https://github.com/reactphp/datagram) purely for
   organizational reasons to avoid a cyclic dependency between the two
--- a/src/Query/CoopExecutor.php
+++ b/src/Query/CoopExecutor.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace React\Dns\Query;
+
+use React\Promise\Promise;
+
+/**
+ * Cooperatively resolves hosts via the given base executor to ensure same query is not run concurrently
+ *
+ * Wraps an existing `ExecutorInterface` to keep tracking of pending queries
+ * and only starts a new query when the same query is not already pending. Once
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -249,6 +249,26 @@ $executor = new RetryExecutor(
...
+Note that this executor is entirely async and as such allows you to execute
+any number of queries concurrently. You should probably limit the number of
+concurrent queries in your application or you're very likely going to face
+rate limitations and bans on the resolver end. For many common applications,
+you may want to avoid sending the same query multiple times when the first
+one is still pending, so you will likely want to use this in combination with
+a `CoopExecutor` like this:
+
+```php
+$executor = new CoopExecutor(
```
kelunik: Maybe `RequestSharingExecutor` / `QuerySharingExecutor` is a better name?

clue: Maybe. I don't have a strong preference for either name, so I'll leave this up to somebody else to decide if somebody has a strong preference and/or some good arguments :+1:

Issue #128, opened 2019-07-07
Roadmap to stable v1.0.0


clue:
Let's face it, this project is *stable* and has been used in production for years :shipit: 

However, we're currently following a v0.X.Y release scheme (http://sentimentalversioning.org/).

We should finally make this explicit and fully adhere to SemVer and release a stable v1.0.0.

To a large extend, a stable v1.0.0 helps making BC breaks more explicit and thus the whole project more reliable from a consumer perspective. This project is actively maintained and has received some major updates in the last years and only has some minor updates planned in the next weeks. Given our current versioning scheme, we'd like to ensure all anticipated BC breaks will be merged before the planned v1.0.0 release.

As such, I've set up a quick roadmap that enlists the major upcoming changes with planned release dates towards a stable v1.0.0 release:

## v0.4.18 :white_check_mark:  

* Released 2019-07-09
* DNS caching fixes

## v1.0.0

* Planned 2019-Q3?
* Drop deprecated APIs
* Mark internal APIs as such
* No other changes planned, this should merely mark the previous release as "stable"

This ticket aims to serve as a basic overview and does not contain every single change. Please also see the milestone links and the CHANGELOG for more details.

Obviously, this roadmap is subject to change and I'll try to keep it updated as we progress. In order to avoid cluttering this, please keep discussion in this ticket to a minimum and consider reaching out to us through new tickets or Twitter etc.

CharlotteDunois:
I'd like to add that TCP support should be included in the stable reason for various reasons (secure queries (TLS over TCP), extended queries (larger than 512 bytes), etc.).

This is something I'd like to work on myself (as long as no one is faster), but this is nothing I've planned for intermediate working on, but probably rather something for somewhere Q3 or Q4.

clue:
@CharlotteDunois I think you're raising an excellent point, so let me try to explain: I agree that this component would not be considered "feature complete" without sending queries over TCP/IP as discussed in #19 and perhaps #80.

The upcoming release should merely mark our APIs as "stable" in the sense that we do not plan to introduce any BC breaks any time soon. This allows consumers of this package to depend on this API for longer term support (LTS).

It's my understanding we should get the LTS release out rather sooner than later and we can always continue working on additional features in future v1.x.0 releases. In fact, I've already prepared the TCP/IP logic and I'm planning to get this shipped the next few weeks :shipit:

CharlotteDunois:
> 
[...]

PR #133, opened 2019-07-10
Avoid garbage references when DNS resolution rejects on legacy PHP <= 5.6

clue:
These garbage references only show up when a DNS lookup rejects with an
Exception on legacy PHP <= 5.6. These issues have been fixed upstream in
the Promise component for current PHP versions with
https://github.com/reactphp/promise/releases/tag/v2.6.0 but this PR
explicitly unsets known garbage references to avoid unexpected memory
growth on legacy PHP versions.

The downstream Socket component implicitly depends on this because its
test suite currently fails. This changeset can be considered a new
feature in this component which fixes a bug in the test suite of a
downstream component.

Accordingly to the roadmap (https://github.com/reactphp/dns/issues/128) we planned to stop the v0.4.x series, but I consider this to be a notable issue that warrants a v0.4.19 release. Once accepted, I will manually merge this into the current `master` branch targeting the upcoming `v1.0.0` milestone.

Builds on top of #125 and #129
Refs #118

### DIFF
--- a/src/Query/CachingExecutor.php
+++ b/src/Query/CachingExecutor.php
@@ -52,10 +52,14 @@ class CachingExecutor implements ExecutorInterface
                         }
                     );
                 }
-            )->then($resolve, $reject);
+            )->then($resolve, function ($e) use ($reject, &$pending) {
+                $reject($e);
+                $pending = null;
+            });
         }, function ($_, $reject) use (&$pending, $query) {
             $reject(new \RuntimeException('DNS query for ' . $query->name . ' has been cancelled'));
             $pending->cancel();
+            $pending = null;
         });
     }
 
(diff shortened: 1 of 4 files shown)


clue:
For the reference: I've verified this actually fixes the build error for the downstream socket component (https://travis-ci.org/reactphp/socket/builds/556729141) by applying this version manually:

```json
# composer.json.patch
{
    "require": {
        "react/dns": "dev-legacy-garbage as 0.4.19"
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/clue/dns"
        }
    ],
    "config": {
        "platform": {
            "php": "5.6.30"
        }
    }
}

```
```
$ cd ~/workspace/reactphp-socket
$ composer update
$ docker run --rm -it -v `pwd`:/data --entrypoint=vendor/bin/phpunit --workdir=/data php:5.6
```

In other words: Once this PR is in and tagged as v0.4.19, the upstream component will build just fine again without any modification :+1:

PR #147, opened 2019-08-09
Reuse TCP/IP socket connection and pipeline outgoing queries

clue:
The TcpTransportExecutor will now keep the TCP/IP socket connection to
the DNS server open for a short period. Additional queries can reuse the
same socket connection and pipeline multiple queries over this single
socket. This avoids creating multiple socket resources for concurrent
requests and implies the TCP/IP connection overhead only applies once
instead of for each query. Among others, this helps avoiding excessive
resources and significantly improves query performance when using
multiple (concurrent) queries.

This implements RFC 7766 as per https://tools.ietf.org/html/rfc7766
Builds on top of #145
Refs #19

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "require": {
         "php": ">=5.3.0",
         "react/cache": "^1.0 || ^0.6 || ^0.5",
-        "react/event-loop": "^1.0 || ^0.5 || ^0.4 || ^0.3.5",
+        "react/event-loop": "^1.0 || ^0.5",
         "react/promise": "^2.1 || ^1.2.1",
         "react/promise-timer": "^1.2"
     },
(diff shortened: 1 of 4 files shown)


clue:
Updated to work around failing test on legacy PHP 5.3 due to partly closed socket resource :shipit:

composer.json
```diff
@@ -19,5 +19,9 @@
     },
     "autoload-dev": {
         "psr-4": { "React\\Tests\\Dns\\": "tests" }
-    }
+    },
+    "config": {
+    "platform": {
+    "php":"5.3.30"
```
jsor: Missing indention

clue: Thanks for spotting, I've removed the affected rows again (this was only a testing leftover) :shipit:

PR #155, opened 2020-02-25
test against php 7.4

sergiy-petrov:


clue:
@sergiy-petrov Thanks for filing this one, I've just filed #156 which adds PHP 7.4 to the test matrix and also fixes two affected tests :+1:

PR #171, opened 2021-02-10
Improve error handling when sending data to DNS server fails (macOS)

clue:
This changeset improves error handling when sending data to the DNS server fails. This is a quite rare error condition but something that can be reproduced when using faulty DNS server entries on macOS (e.g. https://github.com/beyondcode/expose/issues/25#issuecomment-647367304).

Together with @SimonFrings, I'm currently in the process of migrating from Travis to GitHub actions and improving our test matrix. We've noticed a similar error in the (upcoming) PR for the socket component on macOS only, so I figured it's a good idea to address this upstream (here) instead.

Builds on top of #170 and https://github.com/reactphp/stream/pull/157

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,19 @@ jobs:
       - run: vendor/bin/phpunit --coverage-text -c phpunit.xml.legacy
         if: ${{ matrix.php < 7.3 }}
 
+  PHPUnit-macOS:
+    name: PHPUnit (macOS)
+    runs-on: macos-10.15
+    continue-on-error: true
+    steps:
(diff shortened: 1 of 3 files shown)


clue:
The build currently fails due to another flaky test on macOS, seems to be 3/4 chance of it failing. Looks like it could be related to https://github.com/reactphp/stream/pull/150, but after patching this I still get a 1/3 chance of it failing (https://github.com/clue-labs/dns/runs/1873816008?check_suite_focus=true). I'll look into this again no later than tomorrow :shipit:

jsor (approved):

WyriHaximus (approved):

PR #181, opened 2021-06-28
only return valid IPs

remicollet:
On recent linux distro (e.g. Fedora 33+) Network Manager generates /etc/Resolv.conf with interface references

Example:

```
# Generated by NetworkManager
search remirepo.net
nameserver 192.168.1.1
nameserver fe80::7e94:2aff:fe85:5cce%eno1
```

Last address is not valid, but not filtered by loadResolvConfBlocking, so create unusable configuration
Ex: running react/socket test suite

```
There were 8 errors:

1) React\Tests\Socket\ConnectorTest::testConnectorUsesTcpAsDefaultScheme
InvalidArgumentException: Invalid nameserver address given

/usr/share/php/React/Dns/Query/UdpTransportExecutor.php:115
...
```

This trivial fix filter to only return valid IPs

### DIFF
--- a/src/Config/Config.php
+++ b/src/Config/Config.php
@@ -85,7 +85,7 @@ final class Config
         preg_match_all('/^nameserver\s+(\S+)\s*$/m', $contents, $matches);
 
         $config = new self();
-        $config->nameservers = [];
+        $config->nameservers = array();
 
         /* only valid IP */
         foreach($matches[1] as $n) {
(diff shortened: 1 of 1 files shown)


clue:
@remicollet Thank you for looking into this and filing this PR!

I've also looked into this and have just filed #187 to apply a slightly different logic. Instead of ignoring the entire nameserver entry, it will only ignore the IPv6 zone ID, so it would at least still fall back to `fe80::7e94:2aff:fe85:5cce` as a secondary DNS server in your case.

In either case, it will no longer throw an `InvalidArgumentException` and should work just fine by using the primary DNS server first.

remicollet:
nice

PR #185, opened 2021-11-26
Avoid dependency on `ext-filter`

clue:
This simple changeset avoids an unneeded (and undeclared!) dependency on `ext-filter`. The extension is enabled by default, but can explicitly be disabled with `--without-filter`.

The code in question is already covered by the test suite, so the test suite confirms this changeset does not affect behavior or our public API in any way. Accordingly, this is entirely transparent for all consumers of this package, plus now also supports consumers that happen use `--without-filter`.

See https://www.php.net/manual/en/filter.installation.php
Originally reported in https://github.com/leproxy/leproxy/issues/80
Refs https://github.com/reactphp/socket/pull/17

### DIFF
--- a/src/Query/TcpTransportExecutor.php
+++ b/src/Query/TcpTransportExecutor.php
@@ -139,7 +139,7 @@ class TcpTransportExecutor implements ExecutorInterface
         }
 
         $parts = \parse_url((\strpos($nameserver, '://') === false ? 'tcp://' : '') . $nameserver);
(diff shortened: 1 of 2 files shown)


WyriHaximus (approved):

clue:
For future reference only: The error suppression operator is no longer required for PHP 7.3+, see https://3v4l.org/7T1kL

PR #204, opened 2022-08-03
Update test suite to use default loop

clue:
Builds on top of #182 and https://github.com/reactphp/socket/pull/283

PR #213, opened 2023-06-02
Include timeout logic to avoid dependency on reactphp/promise-timer

clue:
This changeset adds timeout logic to avoid the otherwise unneeded dependency on [reactphp/promise-timer](https://github.com/reactphp/promise-timer). There are plans to deprecate the reactphp/promise-stream package as discussed in https://github.com/orgs/reactphp/discussions/475 and in either case I'd like to remove additional dependencies where possible.

In particular, reactphp/dns and reactphp/socket appear to be the only ReactPHP packages that depend on reactphp/promise-timer at the moment. Most consumers of this package will likely install it as a dependency of reactphp/socket, so most consumers will likely continue installing this dependency until it is also removed from reactphp/socket. I have already prepared a similar follow-up PR for reactphp/socket to remove this dependency to ensure most consumers will no longer install reactphp/promise-timer by default.

This changeset does not otherwise affect our public API, so this should be safe to apply. The test suite confirms this has 100% code coverage and does not otherwise affect our APIs.

Note that this changeset does not preclude the discussion in https://github.com/orgs/reactphp/discussions/475, so whether or not or when the package will be deprecated is still up for debate. If you want to explicitly install this dependency, you can still install it like this:

```bash
composer require react/promise-timer
```

Builds on top of #204, #206 and others
See also https://github.com/reactphp/http/pull/482 for similar changes to remove reactphp/promise-stream from reactphp/http

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,12 +29,12 @@
         "php": ">=5.3.0",
         "react/cache": "^1.0 || ^0.6 || ^0.5",
         "react/event-loop": "^1.2",
-        "react/promise": "^3.0 || ^2.7 || ^1.2.1",
-        "react/promise-timer": "^1.9"
+        "react/promise": "^3.0 || ^2.7 || ^1.2.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5 || ^4.8.35",
-        "react/async": "^4 || ^3 || ^2"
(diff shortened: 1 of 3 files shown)


SimonFrings (approved):
Looks good :+1:

WyriHaximus (approved):

PR #214, opened 2023-06-30
Enable support for DNS over TLS (RFC 7858)

### LINKED ISSUE #80: Support DNS over TLS

lucasnetau:
https://tools.ietf.org/html/rfc7858

Reworking of previous PR #200, Closes #80

Implement TLS as a layered class extending the TcpTransportExecutor

* TlsTransportExecutor sets up sane defaults for TLS context options and port number (if not provided) and applies workarounds for ancient PHP bugs.
* TlsTransportExecutor overrides handleWritable to initialise TLS on the connection then passes control to TcpTransportExecutor.
* TcpTransportExecutor::$socket visibility is changed to protected so that TlsTransportExecutor can enable crypto.
* TcpTransportExecutor::readChunk and TcpTransportExecutor::writeChunk (added) visibility is protected to allow TlsTransportExecutor to set workarounds for ancient PHP version bugs.
* TcpTransportExecutor support passing in Stream Context (https://www.php.net/manual/en/context.php) parameters via the query part of the nameserver URI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ easily be used to create a DNS server.
 * [Advanced usage](#advanced-usage)
   * [UdpTransportExecutor](#udptransportexecutor)
   * [TcpTransportExecutor](#tcptransportexecutor)
+    * [DNS over TLS](#dns-over-tls-dot)
   * [SelectiveTransportExecutor](#selectivetransportexecutor)
   * [HostsFileExecutor](#hostsfileexecutor)
 * [Install](#install)
@@ -335,6 +336,27 @@ $executor = new CoopExecutor(
   organizational reasons to avoid a cyclic dependency between the two
   packages. Higher-level components should take advantage of the Socket
   component instead of reimplementing this socket logic from scratch.
+
+#### DNS over TLS (DoT)
+DoT provides secure DNS lookups over Transport Layer Security (TLS).
+The tls:// scheme must be provided when configuring nameservers to 
+enable DoT communication to a TLS supporting DNS server.
+The port 853 is used by default.
+
+```php
+$executor = new TcpTransportExecutor('tls://8.8.8.8');
+````
+
+> Note: To ensure security and privacy, DoT resolvers typically only support
+  TLS 1.2 and above. DoT is not supported on legacy PHP < 5.6 and HHVM
+
+##### TLS Configuration
+[SSL Context parameters](https://www.php.net/manual/en/context.ssl.php) can be set appending passing query parameters to the nameserver URI in the format `wrapper[parameter]=value`.
+
+```php
+// Verify that the 8.8.8.8 resolver's certificate CN matches dns.google
+$executor = new TcpTransportExecutor('tls://8.8.8.8?ssl[peer_name]=dns.google');
+````
 
 ### SelectiveTransportExecutor
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ easily be used to create a DNS server.
 * [Advanced usage](#advanced-usage)
   * [UdpTransportExecutor](#udptransportexecutor)
   * [TcpTransportExecutor](#tcptransportexecutor)
-    * [DNS over TLS](#dns-over-tls-dot)
+  * [TlsTransportExecutor](#tlstransportexecutor)
   * [SelectiveTransportExecutor](#selectivetransportexecutor)
   * [HostsFileExecutor](#hostsfileexecutor)
 * [Install](#install)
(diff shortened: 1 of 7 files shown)


lucasnetau:
@clue @SimonFrings I've refactored this patch so that TlsTransportExecutor is basically a wrapper around TcpTransportExecutor with the TLS components added in to the constructor and handleWrite. Everything else relies on TcpTransportExecutors implementation.

I was prompted to rewrite this after a recent reminder in a comment on a ticket in phpstan (https://github.com/phpstan/phpstan-src/pull/2467#issuecomment-1596121924).

Please accept this for inclusion in reactphp/dns.

lucasnetau:
@clue @SimonFrings any comments?

SimonFrings:
Hey @lucasnetau, my apologies for the delayed response. We're currently swamped with work, making it challenging to dedicate enough time to properly address these tickets. The good news is, that we're currently in the middle of expanding our team in order to get more things done, so you can expect more frequent answers after September.

I'll do my best to prioritize and review this ticket when time permits (though there are several others in the queue 😅).

lucasnetau:
@SimonFrings any movement?

clue (changes requested):
@lucasnetau Again thank you very much for looking into this and filing this high-quality PR!

I think we all agree that having support for DNS over TLS (DoT) would be fantastic and you've clearly put some effort into this. As much as I love the overall idea, I hate I have to critique some of its implementation details (some of which I've clearly contributed to).

In particular, I can see why the current `TlsTransportExecutor` implementation extends the existing `TcpTransportExecutor`. Unfortunately, it looks like an oversight on my end as the `TcpTransportExecutor` should have been marked `final` to discourage inheritance and promote composition instead (see #134, #145 and #148). Accordingly, I'm not sure we would want to merge a changeset that takes advantage of this oversight and I would rather work towards a *cleaner* solution instead.

I understand this will have a noticeable impact on the current implementation, so I'm not exactly sure this is something that can reasonably be changed as part of this PR. That said, at the same time, I hate to suggest closing this one and to ask you to file (yet another) PR.

I'd like to avoid wasting more of your time, so perhaps it makes sense to set up a quick call to discuss implementation approaches in person? If you're interested, you can check my profile to schedule a call.

Thank you for your contribution!

lucasnetau:
This is entirely frustrating process with artificial blocks and hypothetical what-ifs put in place.

The first PR combined TLS and TCP transport together because TLS is just an additional handshake on top of TCP, it makes sense to have them combined as they are tightly coupled (both in protocol since DoT is TLS+TCP, and in the PHP way of needing to construct the TCP connection first then upgrade the crypto to ensure it is async). In that PR I was requested to pull out the TLS components into a seperate class.

This PR went the inheritance route since TLS only needs to setup sane defaults and override the handleWrite to implement the TLS handshake.

Forcing Composition for this functionality is not possible since the TcpExecutor does not expose the raw socket on which to perform the TLS handshake.

The only remaining thing would be to implement an abstract SocketExecutor which is then extended by Udp, Tcp, and Tls executor code which looks like large amount of refactoring and would still see a heap of code duplication between Tcp and Tls since they share 99% of their implementation.

SimonFrings:
@lucasnetau I can understand your frustration with this, but we're not trying to intentionally block this, we just want to introduce this into the project the right way. I still think this is a great feature addition and I really appreciate all the work you put into this, so maybe we can revisit this in the future together.

lucasnetau:
Hi @SimonFrings, I have no intention or wish to continue development two years after I raised the original PR. I've forked and can maintain for my own needs. There have been no tangible alternative implementations offered and two workable solutions have been rejected outright for very loose reasons with no perceivable benefits other than a pursuit for greatness.

SimonFrings:
@lucasnetau Like I said above, I still think this is a high quality PR and we're thankful for all the work you put into this! I think using this in your fork is a completely valid alternative until a similar feature finds its way into the project :+1:

I don't think this conversation here is much related to the topic of adding support for DNS over TLS anymore, so I'll mark these (mine included) as off-topic. Please understand that we don't build these features for ourselves and have to consider the bigger picture as well, as many other users are affected by these changes. To avoid any inconveniences we always try to avoid adding any bugs (by adding tests to our test suite) or foreseeable BC-breaks.

I hope we cross paths again, keep up the good work!

PR #221, opened 2024-02-26
Chore: Drop $loop usages in favor of Loop::get()

bpolaszek:
Still a few tests to fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,6 @@ jobs:
           - 7.3
           - 7.2
           - 7.1
-          - 7.0
-          - 5.6
-          - 5.5
-          - 5.4
-          - 5.3
     steps:
       - uses: actions/checkout@v4
       - uses: shivammathur/setup-php@v2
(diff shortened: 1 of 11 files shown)


SimonFrings:
Hey @bpolaszek, thank you for your pull request, always happy about contributions :+1:

As I also wrote in https://github.com/reactphp/stream/pull/176, changing the minimal supported PHP version to PHP 7.1 is something we want to see for ReactPHP v3, but I think we should use separate PRs for these topics to avoid bloating each pull request in complexity, which also makes them easier to review. In my opinion, we don't need the PHP 7.1 changes to remove the optional `$loop` parameter here.

The rest of the changes are heading in a good direction. I think you can also take a look at https://github.com/clue/reactphp-redis/pull/156 and use this as inspiration (for changes, PR description, commit history, etc.), as we also removed the optional `$loop` parameter in there. Thanks to our consistency across all our projects, we can apply similar changes in here.

Let's focus on either https://github.com/reactphp/stream/pull/176 or this one here first to get a good feeling on what the PR should look like and then use this knowledge for the other pull requests. This way we can avoid doubling the work :+1:

